PR #2, opened 2015-09-02
add more details and relevant links

waldyrious:
Based on https://groups.google.com/d/msg/julia-users/vlLLmSZVeKk/0f-SbEiICAAJ

Also change the level of the section headings (there is usually only one H1, the main title of the document)


Luthaf:
Thank you !

PR #3, opened 2015-09-02
Fix oversight in last commit ("also also")

waldyrious:
Sorry for not proofreading my previous PR carefully enough. I'm taking the opportunity to add a few other fixes, which should compensate this commit's usefulness :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,32 +6,32 @@ This package allows one to use [conda](http://conda.pydata.org/) as a binary pro
 While other binary providers like [Hombrew.jl](https://github.com/JuliaLang/Homebrew.jl),
 [AptGet](https://en.wikipedia.org/wiki/Advanced_Packaging_Tool#apt-get) or
 [WinRPM.jl](https://github.com/JuliaLang/WinRPM.jl) are platform-specific,
-Conda.jl is a cross-platform alternative. It also also be used without administrator rights,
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -6,32 +6,32 @@ This package allows one to use [conda](http://conda.pydata.org/) as a binary pro
...
 `conda` is a package manager which started as the binary package manager
 for the Anaconda Python distribution, but it also provides arbitrary packages.
-Instead of the full Anaconda distribution, Conda.jl uses the miniconda minimalistic Python environment,
+Instead of the full Anaconda distribution, Conda.jl uses the miniconda Python environment,
 which only includes `conda` and its dependencies.
 
 You can install it by running `Pkg.add("Conda")` at the Julia prompt.
 
-## Basic functionalities
+## Basic functionality
```
Luthaf: My english grammar might be faulty, but isn't the plural required here ? We are talking about the functionalities of the Conda module.

waldyrious: Both ways are valid, e.g. Wiktionary [defines](https://en.wiktionary.org/wiki/functionality) "functionality" as both "the ability to perform a task or function" and "set of functions that something is able or equipped to perform". If I'm not mistaken, the singular "functionality" is more common than "functionalities" when describing the second meaning (i.e. a set).

PR #8, opened 2015-09-16
Add bindir function

dhoegh:
Add bindir  function to get scripts in bin dir according to JuliaLang/IJulia/pull/357

### DIFF
--- a/src/Conda.jl
+++ b/src/Conda.jl
@@ -31,6 +31,24 @@ using JSON
 
 "Prefix for installation of all the packages."
 const PREFIX = abspath(dirname(@__FILE__), "..", "deps", "usr")
+@unix_only begin
+    function bindir(package) 
+        packages = _installed_packages_dict()
+        haskey(packages, package) || error("Package not found")
+        joinpath(PREFIX, "pkgs", packages[package][2], "bin")
+    end
+    scriptsdir() = bindir()
+end
+@windows_only function bindir(package) 
+    packages = _installed_packages_dict()
+    haskey(packages, package) || error("Package not found")
+    joinpath(PREFIX, "pkgs", packages[package][2], "Library", "bin")
+end
+@windows_only function scriptsdir(package) 
+    packages = _installed_packages_dict()
+    haskey(packages, package) || error("Package not found")
+    joinpath(PREFIX, "pkgs", packages[package][2], "Scripts")
+end
 
 @unix_only const conda = joinpath(PREFIX, "bin", "conda")
 @windows_only const conda = joinpath(PREFIX, "Scripts", "conda")
@@ -101,20 +119,26 @@ function update()
     end
 end
 
-"List all installed packages as an array."
-function _installed_packages()
+"List all installed packages as an dict of tuples with (version_number, fullname)."
+function  _installed_packages_dict()
     packages = JSON.parse(readall(`$conda list --json`))
-    regex = r"([\w_-]*)-\d.*"
+    # As julia do not accepts xx.yy.zz.rr version number the last part is removed.
+    # see issue https://github.com/JuliaLang/julia/issues/7282 a maximum of three levels is inserted
+    regex = r"^(.*?)-((\d+\.){1,2}\d+)[^-]*-(.*)$"
+    package_dict = Dict{UTF8String, Any}()
     for i in 1:length(packages)
         m = match(regex, packages[i])
         if m != nothing
-            packages[i] = m.captures[1]
+            package_dict[m.captures[1]] = (convert(VersionNumber, m.captures[2]), packages[i])
         else
             error("Failed parsing string: $(packages[i]). Please open an issue!")
         end
     end
-    return packages
+    return package_dict
 end
+
+"List all installed packages as an array."
+_installed_packages() = keys(_installed_packages_dict())
 
 "List all installed packages to standard output."
 function list()
--- a/src/Conda.jl
+++ b/src/Conda.jl
@@ -31,22 +31,33 @@ using JSON
 
 "Prefix for installation of all the packages."
 const PREFIX = abspath(dirname(@__FILE__), "..", "deps", "usr")
-@unix_only begin
-    function bindir(package) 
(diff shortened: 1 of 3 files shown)


Luthaf:
`bindir(package)` should return the binary directory containing the executables installed with the package. That is not what is implemented here. You will need to use the [version](https://github.com/Luthaf/Conda.jl/blob/master/src/Conda.jl#L125) function to do that, and handle the difference between unix and Windows.

Luthaf:
Also, if you can add a test for that function, to ensure future compatibility, that would be nice !

dhoegh:
As seen from the screenshots above i am not convinced it is necessary to do the long version of `bindir(package) = joinpath(PREFIX, "pkgs", version(package), "bin")` are there some specific concerns regarding the difference between `joinpath(PREFIX, "Scripts")` and `joinpath(PREFIX, "pkgs", version(package), "bin")`?

Luthaf:
There are two issue here for me. First, `Scripts` only contains Python scripts, and no other binary executables (like `curl.exe`, ...). These are in `bin`. And second, there is no `bin` folder at `joinpath(PREFIX, "bin")`.

So if `ipython.exe` is also in `joinpath(PREFIX, "pkgs", version(package), "bin")`, this directory will be the best one. If not, we can either have `bindir` returns the two directories (`Scripts` and `bin`), or add a specific function `scripstdir` for the path to `Scripts`.

dhoegh:
`curl.exe` is for me located in `C:\Users\Hoegh\.julia\v0.4\Conda\deps\usr\pkgs\curl-7.43.0-1\Library\bin` while the `IPython.exe` is in `C:\Users\Hoegh\.julia\v0.4\Conda\deps\usr\pkgs\ipython-3.1.0-py27_1\Scripts`. So I do not see where `bindir` should point Do you have any insight?

Luthaf:
> So I do not see where bindir should point Do you have any insight?

The best way to handle that should be to have `bindir("curl")` which returns `[...]\Conda\deps\usr\pkgs\curl-7.43.0-1\Library\bin`, and `scripstdir("ipython")` which gives either `[...]\Conda\deps\usr\pkgs\ipython-3.1.0-py27_1\Scripts` or `[...]\Conda\deps\usr\Scripts`. The long form may be preferable in the case where more than one version is installed.

`scriptsdir` would be the same than `bindir` on unix.

dhoegh:
I have pushed a new functional version. I do not have anymore time to work on this tonight. Maybe the test could be to find the Conda script. Is there any way the version number could be obtained in a different way than I use in `_installed_packages()`

src/Conda.jl
```diff
@@ -31,9 +31,26 @@ using JSON
...
 "Prefix for installation of all the packages."
 const PREFIX = abspath(dirname(@__FILE__), "..", "deps", "usr")
+for f in [:bin_dir, scripts_dir]
+    @eval begin
+    @unix_only function f(package) 
+        packages = _installed_packages()
+        haskey(packages, package) || error("Package not found")
+        joinpath(PREFIX, "pkgs", "bin")
+    end
+end
```
stevengj: I think you want:
```jl
@unix_only function bin_dir(package)
        packages = _installed_packages()
        haskey(packages, package) || error("Package not found")
        joinpath(PREFIX, "pkgs", "bin")
end
@unix_only scripts_dir(package) = bin_dir(package)
```

src/Conda.jl
```diff
@@ -104,16 +121,20 @@ end
...
+            version = split(m.captures[2], ".")
+            # As julia do not accepts xx.yy.zz.rr version number the last part is removed.
+            # see issue https://github.com/JuliaLang/julia/issues/7282
+            package_index[m.captures[1]] = (VersionNumber(join(version[1:min(length(version), 3)], ".")), packages[i])
         else
             error("Failed parsing string: $(packages[i]). Please open an issue!")
         end
     end
-    return packages
+    return package_index
```
stevengj: If you are going to change the return value, that will mess up other calls to `_installed_packages()` in Conda, no?

Maybe rename this function to `_installed_packages_dict` and define `_installed_packages() = keys(_installed_packages_dict())`?

stevengj:
The usual Julia convention is to omit underscores except when demanded by readability.  So maybe `bindir()` and `scriptdir()`?

src/bindeps_conda.jl
```diff
@@ -1,5 +1,5 @@
 # This file contains the necessary ingredients to create a PackageManager for BinDeps
-using BinDeps
+import BinDeps
```
Luthaf: Why is this change necessary ?

Luthaf:
It is starting to look nice ! 

Two more things: 
- Can you add some test in `tests/runtests.jl` ? Checking the `Conda.bindir("curl")` and `Conda.scriptsdir("curl")` output ?
- Can you add (one-line) documentation for the `bindir` and `scriptsdir` functions ?

Also, I think that `haskey(packages, package) || error("Package not found")` should be `haskey(packages, package) || error("Package '$package' not found")`. Giving more context will help when debugging errors.

src/bindeps_conda.jl
dhoegh: Because BinDeps also have a bindir function and hence gives a warning about conflicts with an existing identyfier and therefore i changed it. If it causes any trouble we could change the bindir function name.

Luthaf: Ok, got it. This would be a nice addition to define specialized version of `BinDeps.bindir` for the `Conda.Manager` type. This will be for another PR !

dhoegh:
I have added doc's and two test.

Luthaf:
Waiting for Appveyor to pass, and then I'll merge that.

stevengj:
Probably you should squash the commits before merging.

Luthaf:
@dhoegh if you have time to squash the commits that would be nice, but else I can merge as it is.

dhoegh:
I do not have time for a squash before Sunday.

Luthaf:
I have done it as of 0d2042ed6c73a3b05ddee9bd340188060b5a5fbd

PR #10, opened 2015-09-23
Add PYTHONDIR.

dhoegh:
Add the path to python as a constant. The need were discovered during https://github.com/stevengj/PyCall.jl/pull/191. When merged could you tag version 1.6, so I can continue my quest of widespread adoption of Conda.

### DIFF
--- a/src/Conda.jl
+++ b/src/Conda.jl
@@ -46,6 +46,11 @@ SCRIPTDIR
 "Prefix for the executable files installed with the packages"
 BINDIR
 
+"Returns the directory for python"
+@windows_only const PYTHONDIR = PREFIX
+"Returns the directory for python"
+@unix_only const PYTHONDIR = BINDIR
+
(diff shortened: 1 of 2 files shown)


dhoegh:
I have added a test.

Luthaf:
Will merge as soon as Appveyor tests pass!

> I can continue my quest of widespread adoption of Conda.

Thank you for doing so!

src/Conda.jl
```diff
@@ -46,6 +46,11 @@ SCRIPTDIR
 "Prefix for the executable files installed with the packages"
 BINDIR
 
+"Returns the directory for python"
+@windows_only const PYTHONDIR = PREFIX
```
stevengj: Probably would be cleaner to wrote `const PYTHONDIR = @windows? PREFIX : BINDIR` ... that way, you wouldn't have to repeat the docstring.

PR #14, opened 2015-09-25
lazy installation: run _install_conda() the first time Conda is used

stevengj:
This installs miniconda only the first time Conda is used (e.g. on the first `Conda.add` or similar), rather than in `Pkg.build`.

This allows packages like PyCall to REQUIRE Conda as an optional dependency provider, but doesn't force users to install miniconda unless it is actually needed.

Fixes stevengj/PyCall.jl#198

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -1,8 +1,0 @@
-using Conda
-
-# Let's see if Conda is installed. If not, let's do that first!
-@unix_only is_installed = isexecutable(Conda.conda)
-@windows_only is_installed =  isfile(Conda.conda * ".exe")
-if !is_installed
-    Conda._install_conda()
-end
--- a/src/Conda.jl
+++ b/src/Conda.jl
@@ -89,38 +89,43 @@ function _installer_url()
     return res
 end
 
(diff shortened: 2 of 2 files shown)


Luthaf:
You are too fast! I was answering on the PyCall issue :smile: 

Instead of sticking calls to  `_install_conda` everywhere, how about calling it in `__init__` ? That way, it will be installed on the `using Conda` invocation, not before.

stevengj:
That would work.   I think it is a bit more convenient to be able to put `using Conda` at the top of a file without worrying that will immediately pull in Conda.   `using` statements can't be put into functions and other scopes without calling `eval`, so conditionally `using Conda` is not as flexible as conditionally calling `Conda.add`.

stevengj:
Also, making `using Conda` install miniconda would make a lot of packages probably install miniconda accidentally, because people are used to having `using` statements unconditionally.   Doing it my way would ensure that miniconda is only installed if someone actually needs it.

Luthaf:
Ok, let's go for this!

stevengj:
Great!

PR #16, opened 2015-09-29
Make it possible to request specific version of a Conda package

meggart:
It might be useful to request a specific version of a Conda package in BinDeps. For example, the latest WIndows binaries of libgdal are broken, so we have to fall back to an old version. This PR makes it possible to do 

    provides(Conda.Manager,"libgdal==1.11.2",libgdal)

and the specified version is installed. See also https://github.com/wkearn/RasterIO.jl/pull/11 to see how it could work.

### DIFF
--- a/src/Conda.jl
+++ b/src/Conda.jl
@@ -189,13 +189,32 @@ function search(package::AbstractString)
     return collect(keys(JSON.parse(readall(_set_conda_env(`$conda search $channels $package --json`)))))
 end
 
+"Search a specific version of a package"
+function search(package::AbstractString,ver::AbstractString)
+    _install_conda()
+    channels = additional_channels()
+    ret=JSON.parse(readall(_set_conda_env(`$conda search $channels $package --json`)))
+    out=ASCIIString[]
+    for k in keys(ret)
+      for i in 1:length(ret[k])
+        ret[k][i]["version"]==ver && push!(out,k)
+      end
+    end
+    out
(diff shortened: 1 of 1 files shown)


src/Conda.jl
```diff
@@ -189,13 +189,33 @@ function search(package::AbstractString)
...
+    end
+    out
+end
+
 "Check if a given package exists."
 function exists(package::AbstractString)
-    if package in search(package)
+    if contains(package,"==")
+      pkg,ver=split(package,"==") #Remove version if provided
+      return pkg in search(pkg,ver) ? true : false
```
Luthaf: This should just be `return pkg in search(pkg, ver)`

src/Conda.jl
```diff
@@ -189,13 +189,33 @@ function search(package::AbstractString)
     return collect(keys(JSON.parse(readall(_set_conda_env(`$conda search $channels $package --json`)))))
 end
 
+"Search a specific version of a package"
+function search(package::AbstractString,ver::AbstractString)
+    _install_conda()
+    channels = additional_channels()
+    ret=JSON.parse(readall(_set_conda_env(`$conda search $channels $package --json`)))
+    out=Array(ASCIIString,0)
```
Luthaf: This can be `ASCIIString[]` for consistency with the rest of the code.

meggart:
Ok, updated the commit.

PR #22, opened 2015-10-26
Fixes windows install issue due to installer do not accept quoted paths

dhoegh:
I have added the test to appveyor.


dhoegh:
Appveyor passes tests. Ping me if any changes need to be implemented.

PR #28, opened 2016-01-02
Fix Conda.update failed due to version number conversion

dhoegh:
When "jpeg-8d-vc10_0" package is installed `Conda.update` would fail due to version number conversion in https://github.com/Luthaf/Conda.jl/blob/fc40ebe5488974a7d4e4edce36a238a6ebe03437/src/Conda.jl#L175. It failed because the regex would not match the version number. I have relaxed the error to a warning, as it can be useful to finish even though the version number was not determined correctly. Secondly I have changed how the version is obtained from a regex to a simple split https://github.com/Luthaf/Conda.jl/commit/d15d3148670bb64f9323ba275373cfe5748101ba#diff-b5a035973669724cfc435a3914eae6ebR169 as `conda list --export` separates names and versions with `'='`. `split` can be used because `'='` is disallowed in names and versions, while `'-'` is allowed in names and hence require a reqex to match name and version correctly.

### DIFF
--- a/src/Conda.jl
+++ b/src/Conda.jl
@@ -162,17 +162,20 @@ end
 "List all installed packages as an dict of tuples with (version_number, fullname)."
 function  _installed_packages_dict()
     _install_conda()
-    packages = JSON.parse(readall(_set_conda_env(`$conda list --json`)))
-    # As julia do not accepts xx.yy.zz.rr version number the last part is removed.
-    # see issue https://github.com/JuliaLang/julia/issues/7282 a maximum of three levels is inserted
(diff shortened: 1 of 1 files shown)


Luthaf:
Seems good to me. Thanks !

dhoegh:
@Luthaf could you tag a new version containing this patch. Nice stats for this repo, you posted on twitter :100:

Luthaf:
Oops, I though I had already published a version … I will do it!

Luthaf:
Done, it is tagged as 0.1.9

PR #29, opened 2016-01-26
also run travis on 0.3

tkelman:
still supported according to require file

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ os:
     - osx
     - linux
 julia:
+    - 0.3
     - release
     - nightly
 notifications:


Luthaf:
If the test pass I am ok with adding that, else do you know if the 0.3 release is still used a lot around?

tkelman:
I do still run into cases where supporting it is still necessary, so would like to  breaking things. However it's perfectly acceptable to change your required version to 0.4, bump the minor package version, and do future features only for 0.4 and newer. The old tags will still be used for any 0.3 users.

Luthaf:
Well, test pass for now, so I'll think about it when they will break !

PR #31, opened 2016-05-08
Fix depwarn on 0.5

yuyichao:
Requires JuliaLang/Compat.jl#192 and a new tag.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,4 @@
 julia 0.3
 BinDeps
-Compat
+Compat 0.7.15
 JSON
--- a/src/Conda.jl
+++ b/src/Conda.jl
@@ -29,6 +29,7 @@ provides(Conda.Manager, "libnetcdf", netcdf)
 """
 module Conda
 using Compat
+import Compat.String
 using JSON
 
 "Prefix for installation of all the packages."
@@ -64,7 +65,7 @@ function _set_conda_env(cmd)
     setenv(cmd, env)
 end
 
-const CHANNELS = UTF8String[]
+const CHANNELS = Compat.UTF8String[]
 "Get the list of additional channels"
 function additional_channels()
     res = AbstractString[]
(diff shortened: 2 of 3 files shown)


REQUIRE
```diff
@@ -1,4 +1,4 @@
 julia 0.3
 BinDeps
-Compat
+Compat 0.7.14+
```
Luthaf: Should'nt that line be `Compat 0.7.14` ? The Travis failure seems related to that.

Luthaf:
Travis failure is related to the fact that the Compat changes did not land yet, I am correct?

yuyichao:
Right. This should not be merged until the linked Compat PR is merged and a new version is tagged. The Compat version in `REQUIRE` is now a place holder and needs to be changed to the actual version when the Compat tag happends.

yuyichao:
Updated.

Potential breaking changes:
https://github.com/Luthaf/Conda.jl/pull/31/files#diff-b5a035973669724cfc435a3914eae6ebL170 (if someone depends on the type of the return value)
https://github.com/Luthaf/Conda.jl/pull/31/files#diff-b5a035973669724cfc435a3914eae6ebL222 (if someone depends on the type of the return value)
https://github.com/Luthaf/Conda.jl/pull/31/files#diff-98af5ab944fe38d087a66203b1868a0dL5 (unlikely, only if someone read the field or depend on the aliasing of the arrays.)

yuyichao:
Updated to take advantage of JuliaLang/Compat.jl#198. This doesn't require a new Compat version to work but does require a new tag to get rid of the depwarns on 0.5.

Luthaf:
Thank you for this PR!

I did not follow the recent changes in Julia, the String type is now always UTF-8?

yuyichao:
> the String type is now always UTF-8?

`ASCIIString` is gone.
String literals are always (UTF8)`String`
`UTF8String` is renamed to `String`.

PR #38, opened 2016-07-29
Compat deprecated OS_NAME, OS related macros

wookay:
Fixed deprecated warnings for OS_NAME, OS related macros


Luthaf:
Thank you !

Issue #15, opened 2015-09-26
The silent miniconda setup crashes Outlook on Windows


davidanthoff:
This is really bad, because it can lead to real data loss in Outlook (for example if you have composed a long email but haven't saved/sent it yet). I reported the problem previously at conda/conda#1084, but no reaction there. The issue in the conda repo also points to some discussion of this bug on some mailing list that others have encountered, the root problem seems to be an NSIS problem.

@stevengj This now also affects PyPlot and IJulia, both packages now crash Outlook when you add them (i.e. when the Conda build.jl script runs). When I previously worked on my version of a private Python setup for IJulia, that was the point when everyone agreed that this was unacceptable behavior, and then I worked on a setup that doesn't use miniconda, but a private Python install and used pip instead. I think at this point both IJulia and PyPlot should probably drop the dependency on Conda until this issue is fixed.

stevengj:
As of #14, miniconda now no longer installs on `Pkg.build("Conda")`, but only if Conda is actually needed to install something.   So, people who already have Python are not affected by this.

davidanthoff:
Ah, that is good. Lets hope this can get fixed upstream, it still is a nasty bug.

Luthaf:
> it still is a nasty bug.

Yes it is! I can disable the silent install for now.

stevengj:
If you disable silent install, does that mean that users have to click through some dialogs?  If they make the wrong choices, will it screw up the install?

Luthaf:
I do not know. I'll have to test.

davidanthoff:
Another option might be to test whether the Outlook.exe process is running before the miniconda process is started, and if it is running, just throw an error with a message that users should shutdown Outlook and then repeat the last operation. And then hope that they fix this upstream and then remove this check again.

Checking is probably easiest done using Powershell, the command ``Get-Process Outlook`` should probably do it, I think one would just have to check whether that throws an error (not running) or returns something (running).

stevengj:
There is a [win32 API to get the list of running processes](https://msdn.microsoft.com/en-us/library/windows/desktop/ms682623(v=vs.85).aspx), but I agree that PowerShell seems a lot easier.  

Presumably, this check only needs to be in `Conda._install`.

Luthaf:
How can we call powershell from Julia? I am not a windows expert at all ...

davidanthoff:
I think something like this should work

````julia
try
  run(`powershell "Get-Process Outlook"`)
[...]

PR #43, opened 2016-08-18
fix #25 (new fix for #15 that does not involve powershell)

### LINKED ISSUE #25: Remove message that Outlook process was not found
On Windows I get this:

````
julia> using ExcelReaders
INFO: Precompiling module ExcelReaders...
INFO: Installing xlrd via the Conda package...
Get-Process : Cannot find a process with the name "Outlook". Verify the process name and call the cmdlet again.
At line:1 char:1
+ Get-Process Outlook
+ ~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (Outlook:String) [Get-Process], ProcessCommandException
    + FullyQualifiedErrorId : NoProcessFoundForGivenName,Microsoft.PowerShell.Commands.GetProcessCommand

Fetching package metadata: ....
````

This is on a system with no Outlook installed. The line that starts with ``get-process`` is red, indicating an error. Would be nice if that red output could somehow be suppressed. Not urgent, though, this is more aesthetic.

stevengj:
This fixes the unwanted PowerShell output of #25, by implementing an entirely different fix for #15 that invokes the Win32 API directly rather than forking powershell.

Unfortunately, I had to drop support for Julia 0.3, as the `NTuple` field of `PROCESSENTRY32` does not work in 0.3 (it doesn't inline in the struct), and the alternative of declaring 260 `UInt8` fields explicitly was too horrible to contemplate.

However, at this point in time, with 0.5 just about out the door and 0.3 unbearably ancient, I think dropping 0.3 support seems reasonable.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ os:
     - osx
     - linux
 julia:
-    - 0.3
     - 0.4
     - 0.5
     - nightly
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,4 @@
-julia 0.3.3
+julia 0.4
 BinDeps
-Compat 0.7.20
+Compat 0.8
 JSON
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,7 +1,5 @@
 environment:
   matrix:
-  - JULIAVERSION: "julialang/bin/winnt/x86/0.3/julia-0.3-latest-win32.exe"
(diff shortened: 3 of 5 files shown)


Luthaf:
Seems good to me, thank you!

tkelman:
We should probably create a central WinAPI.jl package to collect various snippets like this.

PR #70, opened 2017-03-24
sanitize PYTHON environment vars when running conda

### LINKED ISSUE #69: sanitize PYTHON* environment variables when running conda
We already remove CONDA_* environment variables when running `conda`, but we should probably remove [PYTHONHOME etcetera too](https://docs.python.org/2/using/cmdline.html)

stevengj:
This fixes #69.  The motivation here is that the `conda` program runs `python` using its own `python` executable, but this can get confused if the user has set PYTHONHOME or similar environment variables because of a different Python installation.

### DIFF
--- a/src/Conda.jl
+++ b/src/Conda.jl
@@ -107,13 +107,13 @@ const CONDARC = conda_rc(ROOTENV)
 """
 Use a cleaned up environment for the command `cmd`.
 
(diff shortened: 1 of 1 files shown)


stevengj:
Travis failure seems to be an unrelated network glitch.

PR #82, opened 2017-07-23
Fix most depwarn on 0.7

yuyichao:
Drop 0.5 support due to Compat.
Also remove 0.4 compat usage that's not needed anymore.


yuyichao:
Ping.

yuyichao:
FYI, I am somehow a member without write access to any JuliaPy repos so someone else need to merge this.

PR #103, opened 2018-03-19
updates for Julia 0.7, version parsing, and latest conda version

stevengj:
This PR includes several updates:

* Fix 0.7 deprecations.

* Fix test failures on latest conda version — previously, each environment got its own `conda` executable, but nowadays it seems that they all use the `conda` in the root environment.  Various other portions of the tests seem to have bitrotted as well.

* Use the VersionParsing package to parse non-semver package version numbers into `VersionNumber`.

* Drops 0.5 support.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ os:
     - osx
     - linux
 julia:
-    - 0.5
     - 0.6
     - nightly
 env:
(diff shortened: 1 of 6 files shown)


stevengj:
Failing to precompile BinDeps on 0.6...

PR #108, opened 2018-07-06
change default to Python 3

stevengj:
The Python ecosystem finally seems to be moving decisively towards Python 3, so I think it makes sense to make Python 3 the default for new Conda.jl installations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,6 @@
 [![Build Status -- OS X and Linux](https://travis-ci.org/JuliaPy/Conda.jl.svg?branch=master)](https://travis-ci.org/JuliaPy/Conda.jl)
 [![Build status -- Windows](https://ci.appveyor.com/api/projects/status/edlxohso05re3v40/branch/master?svg=true)](https://ci.appveyor.com/project/StevenGJohnson/conda-jl)
 
-
 This package allows one to use [conda](http://conda.pydata.org/) as a binary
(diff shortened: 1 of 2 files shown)


stevengj:
@nalimilan, @JobJob, any thoughts on this?

JobJob:
Good idea.

After resisting Python 3 for the better part of a decade, I switched last year. Haven't had any real issues. Seems like the vast majority of packages support Python 3 nowadays.

stevengj:
Done.

Balinus:
Just wondering, is there any way to tell Conda to use python2 ?

ClimateTools.jl uses Basemap as a dependency and there is a lot of problem installing Basemap on Python 3 > 3.6. 

Still haven't found a TravisCI set-up that would solve Basemap problem's with Python 3.7!

Basemap works with Python 3.4 btw.

stevengj:
It’s explained in the Conda README.

Balinus:
Oh!! Sorry about that. Thanks! :+1:

tkf:
`Conda.add("python=3.6"); Pkg.build("PyCall")` may work, too.

oxinabox:
> `Conda.add("python=3.6"); Pkg.build("PyCall")` may work, too.

I found that not to work great.

tkf:
It works on all CIs #129 so it probably is specific to your setting.

PR #113, opened 2018-08-08
Fixes for 1.0

quinnj:
This fixes building Conda.jl on 1.0.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -1,6 +1,7 @@
 using Compat
 
 module DefaultDeps
+    using Compat
     if isfile("deps.jl")
         include("deps.jl")
     end
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -4,10 +4,10 @@ module DefaultDeps
     if isfile("deps.jl")
         include("deps.jl")
     end
-    if !isdefined(:ROOTENV)
+    if !(@isdefined ROOTENV)
         const ROOTENV = abspath(dirname(@__FILE__), "usr")
     end
-    if !isdefined(:MINICONDA_VERSION)
+    if !(@isdefined MINICONDA_VERSION)
         const MINICONDA_VERSION = "3"
     end
 end
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -4,11 +4,20 @@ module DefaultDeps
     if isfile("deps.jl")
         include("deps.jl")
     end
-    if !(@isdefined ROOTENV)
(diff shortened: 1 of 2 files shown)


deps/build.jl
```diff
@@ -32,7 +32,7 @@ const ROOTENV = "$(escape_string(ROOTENV))"
 const MINICONDA_VERSION = "$(escape_string(MINICONDA_VERSION))"
 """
 
-if !isfile("deps.jl") || readstring("deps.jl") != deps
+if !isfile("deps.jl") || String(read("deps.jl")) != deps
```
fredrikekre: `read("deps.jl", String)`?

jekbradbury:
`@isdefined` isn't defined on 0.6, while both REQUIRE and Travis indicate that Conda master is meant to continue working on 0.6, so those lines may need to be guarded with `if VERSION`

devgeist42:
How about
```
module DefaultDeps
    if VERSION < v"0.7"
        macro isdefined(x)
            isdefined(x)
        end
    end
    if isfile("deps.jl")
        include("deps.jl")
    end
    if !(@isdefined ROOTENV)
        const ROOTENV = abspath(dirname(@__FILE__), "usr")
    end
    if !(@isdefined MINICONDA_VERSION)
        const MINICONDA_VERSION = "3"
    end
end
```

jekbradbury:
I went with something similar; will merge if Travis is happy because no IPython on 1.0 is 😢.

deps/build.jl
```diff
@@ -4,10 +4,10 @@ module DefaultDeps
     if isfile("deps.jl")
         include("deps.jl")
     end
-    if !isdefined(:ROOTENV)
+    if !isdefined(@__MODULE__, :ROOTENV)
```
yuyichao: Errr, I think you need Compat... The using above is not in the right module iiuc.

jekbradbury:
The remaining issue is that the Appveyor script uses `julia -e 'versioninfo()`, which needs `using InteractiveUtils` on 1.0. Should Conda switch to the Appveyor.jl version of the script?

stevengj:
Merging this as-is, so that people can use Conda.jl under 1.0.   Fixing AppVeyor can be a separate PR.

stevengj:
Merging this as-is, so that people can use Conda.jl under 1.0.   Fixing AppVeyor can be a separate PR.

fredrikekre:
Can we make a new release such that people can use IJulia on v1.0.

PR #121, opened 2018-09-20
Set PATH before running conda

### LINKED ISSUE #125: [Waiting for upstream] Revert pinning Miniconda3-4.5.4 (Python 3.6) on Windows
This `if` block I introduced in #124 should be removed once conda can install Python 3.7 packages on Windows properly from Anaconda's default channel:

https://github.com/JuliaPy/Conda.jl/blob/7094e2aad2aac5e21eb6db66ebab91c99bad3498/src/Conda.jl#L131-L137

Also, this line should be changed to `Conda.add("python", env)`:

https://github.com/JuliaPy/Conda.jl/blob/7094e2aad2aac5e21eb6db66ebab91c99bad3498/test/runtests.jl#L17

@tkelman Do you know where is the upstream issue we should watch?  Is it https://github.com/conda/conda/issues/7789?

tkf:
close #125

### DIFF
--- a/src/Conda.jl
+++ b/src/Conda.jl
@@ -118,7 +118,7 @@ function _set_path(env_var, env::Environment)
         all_keys = collect(keys(env_var))
         ikey = findfirst(x -> uppercase(x) == "PATH", all_keys)
         has_key = ikey !== nothing
-        path_key = all_keys[ikey]
+        path_key = has_key ? all_keys[ikey] : "PATH"
     else
         path_key = "PATH"
         has_key = haskey(env_var, path_key)
--- a/src/Conda.jl
+++ b/src/Conda.jl
@@ -103,7 +103,53 @@ function _set_conda_env(cmd, env::Environment=ROOTENV)
     env_var["PYTHONIOENCODING"]="UTF-8"
     env_var["CONDARC"] = conda_rc(env)
     env_var["CONDA_PREFIX"] = prefix(env)
+    _set_path(env_var, env)
     setenv(cmd, env_var)
+end
+
+const path_sep = Compat.Sys.iswindows() ? ';' : ':'
+
+function _set_path(env_var, env::Environment)
+    if Compat.Sys.iswindows()
+        # Environment variables are case-insensitive.  `Base.EnvDict`
+        # treats case sensitivity but `env_var` is an arbitrary
+        # dicttionary.  So, we need to treat case-insensitive lookup
+        # ourselves:
+        all_keys = collect(keys(env_var))
+        ikey = findfirst(x -> uppercase(x) == "PATH", all_keys)
+        has_key = ikey !== nothing
+        path_key = all_keys[ikey]
+    else
+        path_key = "PATH"
+        has_key = haskey(env_var, path_key)
+    end
+
+    if has_key
+        env_var[path_key] = extra_path(env) * path_sep * env_var[path_key]
+    else
+        env_var[path_key] = extra_path(env)
+    end
+end
+
+"""
+    extra_path(env::Environment)
+
+Extra `ENV["PATH"]` required for conda environment `env`.
+"""
+function extra_path(env::Environment)
+    if Compat.Sys.iswindows()
+        return join([
+            prefix(env),
+            joinpath(prefix(env), "Library", "mingw-w64", "bin"),
+            joinpath(prefix(env), "Library", "usr", "bin"),
+            joinpath(prefix(env), "Library", "bin"),
+            joinpath(prefix(env), "Scripts"),
+            joinpath(prefix(env), "bin"),
+        ], path_sep)
+        # https://github.com/conda/conda/blob/4.6.0b0/conda/activate.py#L399-L404
+    else
+        return bin_dir(env)
+    end
 end
 
 "Run conda command with environment variables set."
--- a/src/Conda.jl
+++ b/src/Conda.jl
@@ -174,13 +174,6 @@ function _installer_url()
     elseif Compat.Sys.islinux()
         res *= "Linux"
     elseif Compat.Sys.iswindows()
-        if MINICONDA_VERSION == "3"
-            # Quick fix for:
-            # * https://github.com/JuliaLang/IJulia.jl/issues/739
-            # * https://github.com/ContinuumIO/anaconda-issues/issues/10082
-            # * https://github.com/conda/conda/issues/7789
-            res = "https://repo.continuum.io/miniconda/Miniconda$(MINICONDA_VERSION)-4.5.4-"
-        end
         res *= "Windows"
     else
         error("Unsuported OS.")
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -22,6 +22,8 @@ Conda.add("curl", env)
     @test_throws Exception run(cmd)
     Conda.add("pyzmq", env)
     run(cmd)
+
+    Conda.add("jupyter", env)
 end
 
 curlvers = Conda.version("curl",env)
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -3,6 +3,24 @@ using Compat
 using Compat: @info
 using Compat.Test
 
+@testset "_set_path" begin
+    env_var = Dict("PATH" => "somewhere")
+    Conda._set_path(env_var, :dummy)
+    @test endswith(env_var["PATH"], "somewhere")
+    @test occursin("dummy", env_var["PATH"])
+
+    env_var = Dict{String, String}()
+    Conda._set_path(env_var, :dummy)
+    @test occursin("dummy", env_var["PATH"])
+
+    if Compat.Sys.iswindows()
+        env_var = Dict("PaTh" => "somewhere")
+        Conda._set_path(env_var, :dummy)
+        @test endswith(env_var["PaTh"], "somewhere")
+        @test occursin("dummy", env_var["PaTh"])
+    end
+end
+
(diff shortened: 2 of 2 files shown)


src/Conda.jl
```diff
@@ -103,6 +103,8 @@ function _set_conda_env(cmd, env::Environment=ROOTENV)
     env_var["PYTHONIOENCODING"]="UTF-8"
     env_var["CONDARC"] = conda_rc(env)
     env_var["CONDA_PREFIX"] = prefix(env)
+    path_sep = Compat.Sys.iswindows() ? ';' : ':'
+    env_var["PATH"] = bin_dir(env) * path_sep * ENV["PATH"]
```
stevengj: `get(ENV, "PATH", "")`

test/runtests.jl
```diff
@@ -7,6 +10,18 @@ env = :test_conda_jl
 @test Conda.exists("curl", env)
 Conda.add("curl", env)
 
+Conda.add("python", env)
+
+cmd = `python -c "import sys; print(sys.executable)"`
+path = rstrip(read(_set_conda_env(cmd, env), String))
+@test startswith(normpath(path), normpath(bin_dir(env)))
```
stevengj: `startwith` or `normpath(dirname(path)) == normpath(bin_dir(env))`?

test/runtests.jl
```diff
@@ -1,18 +1,32 @@
...
 @test Conda.exists("curl", env)
 Conda.add("curl", env)
 
+@testset "Install Python package" begin
+    Conda.add("python", env)
+    pythonpath = joinpath(Conda.python_dir(env), "python" * exe)
+    @test isfile(pythonpath)
+
+    cmd = Conda._set_conda_env(`$pythonpath -c "import zmq"`, env)
+    @test_throws Exception run(cmd)
```
stevengj: This won't work if `pyzmq` is already installed — people don't necessarily run `test Conda` on a clean installation.

stevengj: In particular, what happens if you run the tests twice in a row?  Won't `pyzmq` still be installed from the previous run?

stevengj:
I'm still unclear why it is necessary to set the `PATH`.  Can't you just run `python` by specifying the full pathname?

test/runtests.jl
tkf: Oops, `@test !success(cmd)` was just a leftover from the previous version.

To make `]test Conda` easier, how about doing

```
try
    Conda.rm("--all", env)
catch
end
```

(or with something similar with cleaner handling of exception)?

tkf:
I don't know, but from https://github.com/JuliaLang/IJulia.jl/issues/739#issuecomment-422989285 and https://github.com/JuliaPy/Conda.jl/issues/119#issuecomment-422253252 (and that `conda` is not _super_ well-behaved package manager, at least ATM), I just thought I'd try.  But making the test in #122 fail properly was already hard...

tkf:
BTW, when I run `conda shell.powershell activate base` (which is, I suppose, the internal of `conda activate`), I get:

```
$env:CONDA_DEFAULT_ENV = "base"
$env:CONDA_EXE = "/home/takafumi/miniconda3/bin/conda"
$env:CONDA_PREFIX = "/home/takafumi/miniconda3"
$env:CONDA_PROMPT_MODIFIER = "(base) "
$env:CONDA_PYTHON_EXE = "/home/takafumi/miniconda3/bin/python"
$env:CONDA_SHLVL = "1"
$env:PATH = "/home/takafumi/miniconda3\bin;..."
```

(the output is weird since I'm on linux)

So it seems `PATH` is the only relevant thing `conda activate ...` alters.

src/Conda.jl
```diff
@@ -103,6 +103,12 @@ function _set_conda_env(cmd, env::Environment=ROOTENV)
     env_var["PYTHONIOENCODING"]="UTF-8"
     env_var["CONDARC"] = conda_rc(env)
     env_var["CONDA_PREFIX"] = prefix(env)
+    path_sep = Compat.Sys.iswindows() ? ';' : ':'
+    if haskey(ENV, "PATH")
+        env_var["PATH"] = bin_dir(env) * path_sep * ENV["PATH"]
```
tkf: https://github.com/JuliaPy/Conda.jl/pull/121#discussion_r219133493: Good catch, but `get(ENV, "PATH", "")` could put `$PWD` into `$PATH`?

tkf:
So it turned out Windows searches `PATH` for a DLL https://msdn.microsoft.com/en-us/library/7d83bc18.aspx?f=255&MSPPError=-2147217396 (via https://github.com/ContinuumIO/anaconda-issues/issues/10082#issuecomment-422834341)

tkf:
But only (random?) half of the tests are fixed by this https://ci.appveyor.com/project/StevenGJohnson/conda-jl/build/1.0.121 So strange...

tkf:
Reading https://github.com/ContinuumIO/anaconda-issues/issues/10082, it looks like it would be fixed on conda side.  Maybe it's better to do `Conda.add("python=3.6")` inside `Conda._install_conda` while waiting for that?

stevengj:
I'm worried that this will also screw up PyCall… importing Python modules won't work if the Python modules depend on the `PATH` being set.

tkf:
Yeah, Conda.jl has to mutate `ENV`. Or maybe PyCall.jl can temporary mutate `ENV` in `pyimport_conda`?  (Similar hack for `LD_LIBRARY_PATH` on Linux won't work, but I don't know what happens in Windows.)

But that's ugly so `Conda.add("python=3.6")` started sound better to me...  Or using slightly old miniconda3 URL would be the easiest solution?

src/Conda.jl
```diff
@@ -103,9 +103,55 @@ function _set_conda_env(cmd, env::Environment=ROOTENV)
...
+        env_var[path_key] = extra_path(env)
+    end
+end
+
+"""
+    extra_path(env::Environment)
+
+Extra `ENV["PATH"]` required for conda environment `env`.
+"""
+function extra_path(env::Environment)
```
tkf: Somebody has to write a code to pass this to SetDllDirectory or AddDllDirectory in `PyCall.__init__`: https://github.com/JuliaPy/Conda.jl/issues/125#issuecomment-423847333

src/Conda.jl
```diff
@@ -103,9 +103,55 @@ function _set_conda_env(cmd, env::Environment=ROOTENV)
...
     env_var["CONDA_PREFIX"] = prefix(env)
+    _set_path(env_var, env)
     setenv(cmd, env_var)
 end
 
+const path_sep = Compat.Sys.iswindows() ? ';' : ':'
+
+function _set_path(env_var, env::Environment)
+    if Compat.Sys.iswindows()
+        # Environment variables are case-insensitive.  `Base.EnvDict`
```
stevengj: Can you file a Julia issue for this?  Seems like it should be fixed in Base.

src/Conda.jl
```diff
@@ -103,9 +103,55 @@ function _set_conda_env(cmd, env::Environment=ROOTENV)
...
+"""
+    extra_path(env::Environment)
+
+Extra `ENV["PATH"]` required for conda environment `env`.
+"""
+function extra_path(env::Environment)
+    if Compat.Sys.iswindows()
+        return join([
+            prefix(env),
+            joinpath(prefix(env), "Library", "mingw-w64", "bin"),
```
stevengj: What if it is 32-bit Windows?

src/Conda.jl
```diff
@@ -128,13 +174,6 @@ function _installer_url()
     elseif Compat.Sys.islinux()
         res *= "Linux"
     elseif Compat.Sys.iswindows()
-        if MINICONDA_VERSION == "3"
-            # Quick fix for:
-            # * https://github.com/JuliaLang/IJulia.jl/issues/739
-            # * https://github.com/ContinuumIO/anaconda-issues/issues/10082
-            # * https://github.com/conda/conda/issues/7789
-            res = "https://repo.continuum.io/miniconda/Miniconda$(MINICONDA_VERSION)-4.5.4-"
-        end
```
stevengj: Seems like this PR could be done before #125.

tkf: Well..., the id of this PR is 121 (< #124)

Joking aside, this PR needs a coordination with PyCall to use SetDllDirectory/AddDllDirectory and for IJulia to tweak setup `ENV["PATH"]` before launching Jupyter Notebook. So I still think it makes sense to release a quick fix with #124 first to help Windows users.

I don't know much about Windows so please feel free to close this PR and implement another solution. I re-opened this PR because I though I'd better cleanup my suboptimal solution.

src/Conda.jl
tkf: It's just because

```
julia> typeof(copy(ENV))
Dict{String,String}
```

To solve this, I think you need to add a case-insensitive in-memory dict to `Base`.

stevengj:
Sorry I haven't looked at this for a while.  CI is failing?

src/Conda.jl
stevengj: Seems like the easiest fix would be to replace `copy(ENV)` with
```
env_var = Dict(uppercase(k)=>v for (k,v) in ENV
```
(Otherwise the other code in `_set_conda_env` would have to be fixed too.)

tkf:
I forgot the status of this PR :)

But I remember thinking that new conda version may provide a better way to do this (via `conda run`).  I'll check if we can do this without hard-coding the paths for Windows.

tkf:
It looks like #170 supersedes this PR. Closing...

PR #126, opened 2018-10-11
install into global ~/.julia/conda directory

### LINKED ISSUE #123: install into a single global directory?
Currently we install into the `deps` directory.  However, in Julia 1.0, packages are no longer updated in place, this means that every time you update Conda.jl it needs to re-install anaconda, which is wasteful and breaks things like PyCall that expect `libpython` to stay in one place.

My thinking is that we should put the root environment into `~/.julia/$JLENV/conda$(CONDA_JL_VERSION)` instead, where `JLENV` is the current Julia package environment.  This way, we will only have a single Anaconda installation (per Python major version) that persists across updates.

However, I must admit that I don't fully grok how Pkg environments work.  From reading the docs, I guess there is a stack of environments.  @StefanKarpinski, is there a way to tell which environment Conda.jl was installed into?  Am I thinking about this in the right way?

stevengj:
Fixes #123.

In the future, we should allow Pkg projects to use a separate Conda virtualenv and set of packages, but that seems like it will have to wait for persistent package options support (JuliaLang/Juleps#38 or similar).

Drops Julia 0.6 since in Julia 0.6 there is already a single global directory (`~/.julia/v0.6/Conda/deps/usr`) and I don't think we want to move it.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -10,11 +10,11 @@ module DefaultDeps
     elseif isfile(Main.condadeps)
         include(Main.condadeps)
     end
-    if !isdefined(@__MODULE__, :ROOTENV)
-        const ROOTENV = Main.condadir
-    end
     if !isdefined(@__MODULE__, :MINICONDA_VERSION)
         const MINICONDA_VERSION = "3"
+    end
+    if !isdefined(@__MODULE__, :ROOTENV)
+        const ROOTENV = joinpath(Main.condadir, MINICONDA_VERSION)
     end
 end
 
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -1,12 +1,17 @@
 using Compat
 
+const condadir = abspath(homedir(), ".julia", "conda")
+const condadeps = joinpath(condadir, "deps.jl")
+
 module DefaultDeps
     using Compat
     if isfile("deps.jl")
         include("deps.jl")
+    elseif isfile(Main.condadeps)
+        include(Main.condadeps)
     end
     if !isdefined(@__MODULE__, :ROOTENV)
-        const ROOTENV = abspath(dirname(@__FILE__), "usr")
+        const ROOTENV = Main.condadir
     end
     if !isdefined(@__MODULE__, :MINICONDA_VERSION)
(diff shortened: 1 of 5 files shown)


deps/build.jl
```diff
@@ -1,16 +1,21 @@
 using Compat
 
+const condadir = abspath(homedir(), ".julia", "conda")
```
tkf: Why not `joinpath(DEPOT_PATH[1], "conda")`? But maybe it's fine since `julia` itself hard-codes `~/.julia/config/startup.jl` https://github.com/JuliaLang/julia/blob/b519147d60683c3a0ed587925bb7221153b64b65/base/client.jl#L288

deps/build.jl
```diff
@@ -1,16 +1,21 @@
...
-    if !isdefined(@__MODULE__, :ROOTENV)
-        const ROOTENV = abspath(dirname(@__FILE__), "usr")
+    elseif isfile(Main.condadeps)
+        include(Main.condadeps)
     end
     if !isdefined(@__MODULE__, :MINICONDA_VERSION)
         const MINICONDA_VERSION = "3"
     end
+    if !isdefined(@__MODULE__, :ROOTENV)
+        const ROOTENV = joinpath(Main.condadir, MINICONDA_VERSION)
```
tkf: I suggested `joinpath(Main.condadir, "base")` and let `conda` control the Python version for each Julia environment https://github.com/JuliaPy/Conda.jl/issues/123#issuecomment-423376252.  But I guess it's OK for a quick fix.

tkf: `conda` command installed in Miniconda3 can install any Python version registered in anaconda repo like 2.7.  So, my thought was that, if you need Python 2 for Julia environment `~/.julia/environments/v1.0-with-py27` then you'd create conda environment `~/.julia/conda/envs/v1.0-with-py27` using `~/.julia/conda/base/bin/conda` (which is interpreted by Python 3).  But since such mechanism does not exist (and not easily implementable without package options), I thought switching Python 2 and 3 by `ENV["CONDA_JL_VERSION"]` is a valid workaround.

PR #129, opened 2018-10-13
Default to Python 3.6 instead of 3.7

### LINKED ISSUE #127: Conda.add may downgrade Python
If a package installed via `Conda.add` is not available for installed Python version, `conda` may downgrade Python:
https://discourse.julialang.org/t/help-understanding-pycall-related-travis-failure/15957

The error is cryptic when it's installed via `pyimport_conda` since the Python installation installed in disk is different from the one loaded in PyCall.

Maybe we should pin Python version and provide a nicer error message pointing to a command that re-builds Conda and PyCall? Or detect change in Python version, automatically run Pkg.build, and then throw an error with message saying that you need to restart Julia?

tkf:
fixes #127

---

Checking with PyCall.jl:

* https://github.com/JuliaPy/PyCall.jl/pull/590
    * https://travis-ci.org/JuliaPy/PyCall.jl/builds/440930322
    * https://ci.appveyor.com/project/StevenGJohnson/pycall-jl-nu3aa/builds/19480780

### DIFF
--- a/src/Conda.jl
+++ b/src/Conda.jl
@@ -128,13 +128,6 @@ function _installer_url()
     elseif Compat.Sys.islinux()
         res *= "Linux"
     elseif Compat.Sys.iswindows()
-        if MINICONDA_VERSION == "3"
-            # Quick fix for:
-            # * https://github.com/JuliaLang/IJulia.jl/issues/739
-            # * https://github.com/ContinuumIO/anaconda-issues/issues/10082
-            # * https://github.com/conda/conda/issues/7789
-            res = "https://repo.continuum.io/miniconda/Miniconda$(MINICONDA_VERSION)-4.5.4-"
-        end
         res *= "Windows"
     else
         error("Unsuported OS.")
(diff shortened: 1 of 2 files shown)


stevengj:
If we do this, what is the upgrade path for existing users when we eventually switch to 3.7?  We don't want to tell everyone to `Conda.free("python=3.6")` manually…

stevengj:
One option would be to use `MINICONDA_VERSION="3.6"`, since this string is now saved between updates … if we notice that this number has increased we could free the old version.

tkf:
The easiest (non?) solution is to just downgrade Python to 3.6 without `pin`ning.  It's then almost equivalent to the current situation before this PR except that it reduces the probability of accidental downgrade.  Python would be automatically upgraded by `Conda.add(package)` once 3.6 becomes too old to install the `package`.

If you want a more controlled solution then I guess we can check Python version in `build.jl` or `__init__`?  Automatic upgrade in `build.jl` is probably rather too brutal but maybe we can print something like "Your Python version is becoming old.  Run `Codna.update()` to upgrade Python and Python packages."

tkf:
I think https://github.com/JuliaPy/PyCall.jl/pull/613 is a better solution. It may still make sense to add `Conda.pin` and `Conda.free`.

PR #130, opened 2018-11-08
Allow multiple packages to be added/removed in a batch

tkf:
While writing https://github.com/JuliaPy/PyCall.jl/pull/613 I noticed that this is not supported.

As `conda` command takes sometime to resolve dependencies, I think it makes sense to (un)install packages in a batch.


stevengj:
Just merged this, but in hindsight it would have been good to add a test.

PR #131, opened 2018-11-08
Test batch (un)install

tkf:
Re https://github.com/JuliaPy/Conda.jl/pull/130#issuecomment-436990972. Yes, I should've added tests.


stevengj:
zip and wget aren't available on Windows; try some small pure-Python packages.  )e.g. `affine` and `ansi2html`?)

tkf:
I changed them to affine and ansi2html. Appveyor is green now.

PR #132, opened 2018-12-01
Recreating environments

rofinn:
Added Conda.freeze and Conda.create functions for re-creating environments.

NOTE: I would have done `Conda.export` if `export` wasn't already a keyword, but `freeze` (e.g., `pip freeze`) seemed like the next best thing.

### DIFF
--- a/src/Conda.jl
+++ b/src/Conda.jl
@@ -231,8 +231,12 @@ NOTE: Because export is a reserved word we use the term freeze from python
 function freeze(filepath::AbstractString, env::Environment=ROOTENV)
     _install_conda(env)
     open(filepath, "w") do fobj
-        write(fobj, read(_set_conda_env(`$conda list --export`, env)))
+        freeze(fobj, env)
     end
+end
+
+function freeze(io::IO, env::Environment=ROOTENV)
+    write(io, read(_set_conda_env(`$conda list --export`, env)))
 end
 
 "Get the exact version of a package as a `VersionNumber`."
@@ -340,6 +344,14 @@ function create(
         `$conda create $(_quiet()) -y -p $(prefix(env)) $channel_str --file $filepath`,
         env
     ))
+end
+
+function create(io::IO, args...; kwargs...)
+    mktemp() do path, fobj
+        write(fobj, read(io))
+        close(fobj)
+        create(path, args...; kwargs...)
+    end
 end
 
 end
--- a/src/Conda.jl
+++ b/src/Conda.jl
@@ -225,6 +225,9 @@ function list(env::Environment=ROOTENV)
 end
 
 """
+    export_list(filepath, env=$ROOTENV)
+    export_list(io, env=$ROOTENV)
+
 List all packages and write them to an export file for use the Conda.create
 NOTE: Because export is a reserved word we use the term freeze from python
 """
@@ -333,7 +336,12 @@ function clean(;
     run(_set_conda_env(cmd))
 end
 
-"Create a new environment with various channels and a packages file."
+""""
+    create(filename, env=$ROOTENV, channels=String[])
+    create(io, env=$ROOTENV, channels=String[])
+
+Create a new environment with various channels and a packages file.
+"""
 function create(
     filepath::AbstractString,
     env::Environment=ROOTENV;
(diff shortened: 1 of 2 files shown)


StefanKarpinski:
@stevengj, would you mind taking a look at this?

rofinn:
> Maybe we should use similar terminology to Pkg3, where this is called a "manifest". e.g. Conda.manifest_export(f, ...) and Conda.manifest_import(f, ...).

I think I'd prefer to keep the terminology between python & julia package management separate to avoid confusion (I don't believe the files are referred to as manifests anywhere in the python documentation).

> For the f argument, it would be nice to support either a filename or an arbitrary ::IO object. This is certainly possible for export. For import I guess you'd have to write a temporary file first.

That's true, but I'm not sure what the use case would be. Are you thinking in case someone wants to codify the requirements in a julia string or something?

src/Conda.jl
```diff
@@ -224,6 +224,17 @@ function list(env::Environment=ROOTENV)
...
 end
 
+"""
+List all packages and write them to an export file for use the Conda.create
+NOTE: Because export is a reserved word we use the term freeze from python
+"""
+function freeze(filepath::AbstractString, env::Environment=ROOTENV)
+    _install_conda(env)
+    open(filepath, "w") do fobj
+        write(fobj, read(_set_conda_env(`$conda list --export`, env), String))
```
omus: I think you can just use `read` instead of `read(..., String)` here.

rofinn: Oh, yeah, I guess we're just dumping the output directly to a file anyways.

stevengj:
Yes.  In general, in Julia wherever you can pass a filename you can also pass an `IO` object.

stevengj:
Whatever we call it, I think the names for import and export should echo one another: `foo_import` and `foo_export` or similar.  And the term "manifest" is pretty common in many packaging systems.

I don't like the term `freeze` since "freezing" the package system could also be interpreted as locking it to prevent changes.

rofinn:
Hmmm, if julia had native filepath types then `freeze` could just be a `list` method.

src/Conda.jl
```diff
@@ -224,6 +224,21 @@ function list(env::Environment=ROOTENV)
     runconda(`list`, env)
 end
 
+"""
+List all packages and write them to an export file for use the Conda.create
```
stevengj: The standard docstring format begins with the function signature.  See https://docs.julialang.org/en/v1/manual/documentation/

rofinn: Sure. I was just trying to follow the convention already used in this file. https://github.com/JuliaPy/Conda.jl/blob/master/src/Conda.jl#L92

src/Conda.jl
```diff
@@ -289,4 +307,30 @@ function rm_channel(channel::AbstractString, env::Environment=ROOTENV)
     runconda(`config --remove channels $channel --force`, env)
 end
 
+""""
+    create(filename, env=$ROOTENV, channels=String[])
+    create(io, env=$ROOTENV, channels=String[])
```
stevengj: I still think the name here should mirror the export name.  `import_list` since we have `export_list`

src/Conda.jl
```diff
@@ -224,6 +224,24 @@ function list(env::Environment=ROOTENV)
     runconda(`list`, env)
 end
 
+"""
+    export_list(filepath, env=$ROOTENV)
+    export_list(io, env=$ROOTENV)
+
+List all packages and write them to an export file for use the Conda.create
+NOTE: Because export is a reserved word we use the term freeze from python
```
stevengj: delete

stevengj:
(needs a rebase/merge)

src/Conda.jl
rofinn: That doesn't match up to the conda commands, but sure.

stevengj:
appveyor failure?

stevengj:
There is an error:
```
      File "C:\Users\appveyor\.julia\conda\3\lib\site-packages\conda\models\records.py", line 102, in box
        val = tuple(f for f in (ff.strip() for ff in val) if f)
    TypeError: 'NoneType' object is not iterable
```
that doesn't look like a network problem?

stevengj:
I restarted AppVeyor just in case it was a transient error.

stevengj:
Passing now, so must have been a temporary network glitch.

rofinn:
Thanks!

PR #135, opened 2018-12-17
Conda.cleanall()

rofinn:
Added a Conda.cleanall function for removing unused packages and artifacts.

### DIFF
--- a/src/Conda.jl
+++ b/src/Conda.jl
@@ -289,4 +289,39 @@ function rm_channel(channel::AbstractString, env::Environment=ROOTENV)
     runconda(`config --remove channels $channel --force`, env)
 end
 
+"Cleanup all unused packages and caches"
+function cleanall(; debug=false)
+    clean(; debug=debug, index=true, locks=true, tarballs=true, packages=true, sources=true)
+end
+
+"""
+    clean(;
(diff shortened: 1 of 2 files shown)


src/Conda.jl
```diff
@@ -289,4 +289,25 @@ function rm_channel(channel::AbstractString, env::Environment=ROOTENV)
     runconda(`config --remove channels $channel --force`, env)
 end
 
+"Cleanup all unused packages and caches"
+function cleanall(; debug=false)
+    clean(; debug=debug, index=true, locks=true, tarballs=true, packages=true, sources=true)
+end
+
+function clean(;
```
stevengj: Needs a docstring.  What does this do with the default arguments?

rofinn: Nothing. I suppose we could throw a warning in that condition.

stevengj: If `clean()` does nothing, why not make it do something useful: just make it do`cleanall()` then, and get rid of the `cleanall` function?

rofinn: So have the default kwargs be `true`? That seemed like a dangerous default, but I'm personally always doing `cleanall` anyways.

stevengj: Yes.

PR #139, opened 2019-01-26
Specify channel when adding package

rofinn:
First step in changing the behaviour of `pyimport_conda` to guarantee the installation of a package from a specified channel(e.g., `conda install -y -c conda-forge requests`).


stevengj:
Thanks!

PR #152, opened 2019-04-28
Add Project.toml

tkf:
This patch adds `Project.toml` generated by Pkg.jl's `gen_project.jl` script
and removes `REQUIRE` file.


stevengj:
Do we still need Compat or can that be removed?

tkf:
Yes, it looks like it.

Issue #81, opened 2017-07-18
Conda.add("numpy") fails on 64-bit Windows 10


Asdow:
Hello. I was updating my Julia to 0.6 stable but ever since doing that I can't get PyCall to work and I've tracked it down to Conda not working right. First I got stuck on the endless miniconda install loop. After using the workaround in https://github.com/JuliaPy/Conda.jl/issues/79 I managed to install it, but I still can't get it to work. 
Doing 

```
using Conda

Conda.add("numpy")
```

gives this error
```
Traceback (most recent call last):
  File "I:\Julia\Julia_pkgs\v0.6\Conda\deps\usr\Scripts\conda-script.py", line 5, in <module>
    sys.exit(conda.cli.main())
  File "I:\Julia\Julia_pkgs\v0.6\Conda\deps\usr\lib\site-packages\conda\cli\main.py", line 179, in main
    return conda_exception_handler(_main, *args)
  File "I:\Julia\Julia_pkgs\v0.6\Conda\deps\usr\lib\site-packages\conda\exceptions.py", line 636, in conda_exception_handler
    return handle_exception(e)
  File "I:\Julia\Julia_pkgs\v0.6\Conda\deps\usr\lib\site-packages\conda\exceptions.py", line 626, in handle_exception
    print_unexpected_error_message(e)
  File "I:\Julia\Julia_pkgs\v0.6\Conda\deps\usr\lib\site-packages\conda\exceptions.py", line 588, in print_unexpected_error_message
    stderrlogger.info(get_main_info_str(get_info_dict()))
  File "I:\Julia\Julia_pkgs\v0.6\Conda\deps\usr\lib\site-packages\conda\cli\main_info.py", line 162, in get_info_dict
    from ..connection import user_agent
  File "I:\Julia\Julia_pkgs\v0.6\Conda\deps\usr\lib\site-packages\conda\connection.py", line 12, in <module>
    from requests import Session, __version__ as REQUESTS_VERSION
  File "I:\Julia\Julia_pkgs\v0.6\Conda\deps\usr\lib\site-packages\requests\__init__.py", line 52, in <module>
    from .packages.urllib3.contrib import pyopenssl
  File "I:\Julia\Julia_pkgs\v0.6\Conda\deps\usr\lib\site-packages\requests\packages\urllib3\contrib\pyopenssl.py", line 47, in <module>
    from cryptography import x509
[...]

PR #154, opened 2019-05-14
Instructions for non-ASCII home directories (#81)

simonbyrne:
address #81.


stevengj:
Actually, non-ASCII home directories are supported.  But there is a bug that could arise if the user had an old Conda installation.

PR #162, opened 2019-12-10
Add support for Power9 and error on unsupported Sys.ARCH

ali-ramadhan:
I'm starting to use Julia on Power9 (I believe still officially unsupported) and noticed that Conda.jl downloads and tries to install x86 miniconda on Power9 which fails (see link to sample stacktrace below).

Turns out miniconda provides Power9 builds so this PR adds support for Conda.jl on Power9. Conda.jl should now error when trying to install miniconda on an unsupported architecture.

Could have probably done `res *= string(Sys.ARCH)` but that probably won't work for all `Sys.ARCH`.

Unfortunately I'm not aware of any way of testing this on Travis CI or Appveyor but I ran the Conda.jl tests on a Power9 system and they all passed (see link to log below).

Conda.jl fails on Power9 stacktrace: https://gist.github.com/ali-ramadhan/6a36885905a3f245e7dedb418a867116
Conda.jl test log on Power9: https://gist.github.com/ali-ramadhan/cda0a74931f874040a88fb0cef638754

### DIFF
--- a/src/Conda.jl
+++ b/src/Conda.jl
@@ -145,7 +145,7 @@ function _installer_url()
     elseif Sys.ARCH == :ppc64le
         res *= "-ppc64le"
     else
-        error("Unsupported architecture: $Sys.ARCH")
+        error("Unsupported architecture: $(Sys.ARCH)")
     end
 
     res *= Sys.iswindows() ? ".exe" : ".sh"
--- a/src/Conda.jl
+++ b/src/Conda.jl
@@ -140,10 +140,8 @@ function _installer_url()
         error("Unsuported OS.")
     end
 
-    if Sys.ARCH == :x86 || Sys.ARCH == :x86_64
-        res *= Sys.WORD_SIZE == 64 ? "-x86_64" : "-x86"
-    elseif Sys.ARCH == :ppc64le
-        res *= "-ppc64le"
+    if Sys.ARCH in (:x86, :x86_64, :ppc64le)
+        res *= string('-',Sys.ARCH)
     else
         error("Unsupported architecture: $(Sys.ARCH)")
     end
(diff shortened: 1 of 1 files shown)


christophernhill:
@ali-ramadhan I asked at IBM and they mentioned this 

 https://blog.travis-ci.com/2019-11-12-multi-cpu-architecture-ibm-power-ibm-z . 

It sounds like Travis may have introduced some Power 9 support as of about three week ago.

stevengj:
Since Power9 is experimental for Julia right now, I think it is fine to merge this without testing.

ali-ramadhan:
Hmmm, looks like Appveyor failed on x86 platforms because `Unsupported architecture: i686`. Previously `i686` seemed to map to `x86`. Should `i686` be treated as an edge case?

Appveyor also failed on x64 platforms. Can't say I fully understand the stacktrace though.

vchuravy:
i686 is one of the x86 architectures... and no-one agrees on the naming scheme so some projects use i686 (i.e. the first of the processor line they still support) whereas other people just use x86...

So you need to normalize to whatever the other side expects.

ali-ramadhan:
Are the Appveyor tests the only thing blocking this PR?

It looks like Windows tests have been failing the same test for the past 4 months (https://ci.appveyor.com/project/StevenGJohnson/conda-jl/history) so I don't think this PR caused the failure. 

It looks like certain packages install fine on Windows but pyzmq does not for some reason...

stevengj:
Hmm, looks like Anaconda might need some `PATH` additions on Windows, as described here: https://github.com/nteract/hydrogen/issues/1517#issuecomment-509395721

ali-ramadhan:
Awesome, thank you! Is there any chance we can get a patch release tagged or is the next release scheduled to coincide with Julia 1.4?

stevengj:
We can get a patch release; it’s not synchronized with Julia releases

PR #164, opened 2019-12-20
drop 0.7 support, add Project.toml

stevengj:


stevengj:
whoops, we already have project.toml

PR #167, opened 2020-02-06
Fixing exists by adding keywords

dsweber2:
Exists doesn't work when you have an environment that's a string, such as if you have a different root environment. The first commit just addresses exists, while the second one adds keyword values inspired by https://github.com/JuliaPy/Conda.jl/issues/118.

### DIFF
--- a/src/Conda.jl
+++ b/src/Conda.jl
@@ -291,7 +291,7 @@ function exists(package::AbstractString, env::Environment=ROOTENV)
       pkg,ver=split(package,"==")  # Remove version if provided
       return pkg in search(pkg,ver,env)
     else
-      if package in search(package,env)
+      if package in search(package,Symbol(env))
         # Found exactly this package
         return true
       else
--- a/src/Conda.jl
+++ b/src/Conda.jl
@@ -192,7 +192,11 @@ end
 const PkgOrPkgs = Union{AbstractString, AbstractVector{<: AbstractString}}
 
 "Install a new package or packages."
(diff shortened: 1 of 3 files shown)


src/Conda.jl
```diff
@@ -192,7 +192,11 @@ end
 const PkgOrPkgs = Union{AbstractString, AbstractVector{<: AbstractString}}
 
 "Install a new package or packages."
-function add(pkg::PkgOrPkgs, env::Environment=ROOTENV; channel::AbstractString="")
+function add(pkg::PkgOrPkgs; env::Environment=ROOTENV, channel::AbstractString="")
```
stevengj: This is a breaking change.  You should add:
```jl
@deprecate add(pkg::PkgOrPkgs, env::Environment) add(pkg, env=env)
```
and similar so that callers using the old API still function, albeit with a warning message.

src/Conda.jl
```diff
@@ -192,7 +192,11 @@ end
...
+    add(pkg,env,channel)
+end
+
+function add(pkg::PkgOrPkgs, env::Environment, channel::AbstractString="")
```
stevengj: Why is this 3-argument version here?

stevengj: We should just have the keyword version, plus a deprecated 2-arg version, if we want to do this.

fatteneder:
Hi.
I ran into this issue today. May I ask what is the status of the proposed fix?

PR #168, opened 2020-02-08
_set_conda_env should preserve any existing cmd.env

stevengj:
`_set_conda_env` should preserve any existing environment in the passed `cmd` object.  (This is so that we can use it from IJulia to get the right environment for invoking `jupyter`.)

Also works around JuliaLang/julia#29334 for julia < 1.2


stevengj:
No, this doesn't work because `cmd.env` is not a dictionary.

stevengj:
I used `withenv` instead in JuliaLang/IJulia.jl#900

PR #200, opened 2021-03-23
Friendly error message for users with special characters in user names on Windows

### LINKED ISSUE #199: installation failled with accent in user name
I am trying to help two students who failed to install Conda.jl. A manual miniconda installation to the directory `c:\users\<username>\.julia\conda\3\` also failed. We see the following error in the miniconda installation log:

```
Traceback (most recent call last):
   File "concurrent\futures\process.py", line 368, in
_queue_management_worker
   File "multiprocessing\connection.py", line 251, in recv
TypeError: __init__() missing 1 required positional argument: 'msg'
```

Both students have a French accent in their username which seems to be an issue for conda (at least [anaconda](https://docs.anaconda.com/anaconda/user-guide/faq/)):

> We recommend installing Anaconda or Miniconda into a directory that contains only 7-bit ASCII characters and no spaces, such as C:\anaconda. Do not install into paths that contain spaces such as C:\Program Files or that include Unicode characters outside the 7-bit ASCII character set. This helps ensure correct operation and no errors when using any open-source tools in either Python 3 or Python 2 conda environments.

Could `Conda.jl` detect this issue and propose a work-around to the users?
I cannot replicate the issue locally, but after reading [build.jl](https://github.com/JuliaPy/Conda.jl/blob/e5444efa5d830c6e85f37e61775d6f19f91f2f8c/deps/build.jl) it seems that setting `ENV["CONDA_JL_HOME"]`  might be sufficient. 

See also https://github.com/conda/conda/issues/10239

Alexander-Barth:
This PR closes #199

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,6 +145,30 @@ pkg> build Conda
 Note that Conda.jl 1.6 and above will use miniforge by default on x86_64, aarch64
 and ppc64le systems.
 
+
+## Troubleshooting
+
+### Installation on Windows with special characters in user names
+
+If you have a special character in your user name (like an umlaut, an accent or a space) the installation which defaults to
+directory `C:\Users\<username>\.julia\Conda\3` will fail on Windows.
+This is a [known issue](https://github.com/conda/conda/issues/10239). The work-around is to install Miniconda to a user-writable directory outside of the home directory.
+If you have an unsuccessful installation of `Conda.jl` and/or related packages like `IJulia` or `PyPlot`, it might be necessary to remove these packages first.
+Before installing `Conda.jl`, choose a directory without space and without special characters and set the environment variable `CONDA_JL_HOME` as follows inside a julia session:
+
+```julia
+ENV["CONDA_JL_HOME"] = "C:\\Conda-julia\\3"
+
+# Install Conda.jl
+using Pkg
+Pkg.add("Conda")
+
+using Conda
+# this should be your chosen Conda home directory
+@show Conda.ROOTENV
+```
+
+
 ## Bugs and suggestions
 
 Conda has been tested on Linux, OS X, and Windows.
--- a/README.md
+++ b/README.md
@@ -148,24 +148,32 @@ and ppc64le systems.
 
 ## Troubleshooting
 
-### Installation on Windows with special characters in user names
+### Installation with special characters in user names
 
-If you have a special character in your user name (like an umlaut, an accent or a space) the installation which defaults to
-directory `C:\Users\<username>\.julia\Conda\3` will fail on Windows.
+If you have a special character in your user name (like an umlaut or an accent) the installation which defaults to
+directory `C:\Users\<username>\.julia\Conda\3` will fail on Windows. A space in your user name will also fail on any platform.
 This is a [known issue](https://github.com/conda/conda/issues/10239). The work-around is to install Miniconda to a user-writable directory outside of the home directory.
-If you have an unsuccessful installation of `Conda.jl` and/or related packages like `IJulia` or `PyPlot`, it might be necessary to remove these packages first.
 Before installing `Conda.jl`, choose a directory without space and without special characters and set the environment variable `CONDA_JL_HOME` as follows inside a julia session:
 
 ```julia
-ENV["CONDA_JL_HOME"] = "C:\\Conda-julia\\3"
-
-# Install Conda.jl
+ENV["CONDA_JL_HOME"] = raw"C:\Conda-Julia\3"
 using Pkg
-Pkg.add("Conda")
+Pkg.build("Conda")
+```
 
+After restarting Julia, you can verify the new installation directory:
+
+```julia
 using Conda
-# this should be your chosen Conda home directory
 @show Conda.ROOTENV
+```
+
+If you use `IJulia` or `PyCall`, they need to be re-build:
+
+```julia
+using Pkg
+Pkg.build("PyCall")
+Pkg.build("IJulia")
 ```
 
 
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -55,11 +55,11 @@ USE_MINIFORGE = lowercase(get(ENV, "CONDA_JL_USE_MINIFORGE", DefaultDeps.USE_MIN
 
 if isdir(ROOTENV) && MINICONDA_VERSION != DefaultDeps.MINICONDA_VERSION
     error("""Miniconda version changed, since last build.
-However, a root enviroment already exists at $(ROOTENV).
-Setting Miniconda version is not supported for existing root enviroments.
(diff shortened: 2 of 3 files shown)


deps/build.jl
```diff
@@ -34,16 +34,30 @@ USE_MINIFORGE = lowercase(get(ENV, "CONDA_JL_USE_MINIFORGE", DefaultDeps.USE_MIN
...
+To leave Miniconda version as, it is unset the CONDA_JL_VERSION environment variable and rebuild.
+To change Miniconda version, you must delete the root environment and rebuild.
+WARNING: deleting the root environment will delete all the packages in it.
 This will break many Julia packages that have used Conda to install their dependancies.
 These will require rebuilding.
 """)
 end
 
+
+if Sys.iswindows() && (!isascii(ROOTENV) || ' ' ∈ ROOTENV)
```
stevengj, line 48: This shouldn't go here, it should go into `Conda._install_conda`, so that you only get an error (or a warning?) if you actually try to install Anaconda.

Alexander-Barth:
I placed the error message `build.jl` to prevent the creation of `deps.jl` with an unusable `ROOTENV` and I chose an error (instead of a warning) because the current miniconda installer failed in all cases that I am aware of. But I did not try miniforge or any non-default locale settings/file systems. I have no strong opinion about this.

deps/build.jl
Alexander-Barth: I agree, I updated the PR.

src/Conda.jl
```diff
@@ -189,6 +189,21 @@ _quiet() = get(ENV, "CI", "false") == "true" ? `-q` : ``
 "Install miniconda if it hasn't been installed yet; _install_conda(true) installs Conda even if it has already been installed."
 function _install_conda(env::Environment, force::Bool=false)
     if force || !isfile(Conda.conda)
+        if Sys.iswindows() && (!isascii(PREFIX) || ' ' ∈ PREFIX)
```
stevengj, line 192: Note that spaces in the path are not supported on any operating system, so it might be good to give an error in general in that case.

src/Conda.jl
```diff
@@ -189,6 +189,21 @@ _quiet() = get(ENV, "CI", "false") == "true" ? `-q` : ``
...
+            error("""Conda.jl cannot be installed to its default location $(PREFIX) as
+Miniconda does not support the installation to a directory with a non-ASCII
+character or a space on Windows. The work-around is to install Miniconda to a
+user-writable directory by setting the CONDA_JL_HOME environment
+variable before installing Conda.jl. For example:
+
+ENV["CONDA_JL_HOME"] = "C:\\\\Conda-julia\\\\3"
```
stevengj, line 199: Probably clearer to use `raw"C:\\Conda-Julia\3"`

stevengj: (`\\\\` gives a double backslash; not sure why that would be needed?)

README.md
```diff
@@ -133,6 +133,30 @@ pkg> build Conda
...
 
+
+## Troubleshooting
+
+### Installation on Windows with special characters in user names
+
+If you have a special character in your user name (like an umlaut, an accent or a space) the installation which defaults to
+directory `C:\Users\<username>\.julia\Conda\3` will fail on Windows.
+This is a [known issue](https://github.com/conda/conda/issues/10239). The work-around is to install Miniconda to a user-writable directory outside of the home directory.
+If you have an unsuccessful installation of `Conda.jl` and/or related packages like `IJulia` or `PyPlot`, it might be necessary to remove these packages first.
```
stevengj, line 144: That shouldn't be necessary.

README.md
```diff
@@ -133,6 +133,30 @@ pkg> build Conda
...
+Before installing `Conda.jl`, choose a directory without space and without special characters and set the environment variable `CONDA_JL_HOME` as follows inside a julia session:
+
+```julia
+ENV["CONDA_JL_HOME"] = "C:\\Conda-julia\\3"
+
+# Install Conda.jl
+using Pkg
+Pkg.add("Conda")
```
stevengj, line 152: If they have already added Conda (which seems likely if they are troubleshooting), they should use `Pkg.build` here.

README.md
stevengj: (It should be sufficient to re-build IJulia and PyCall.)

src/Conda.jl
```diff
@@ -189,6 +189,21 @@ _quiet() = get(ENV, "CI", "false") == "true" ? `-q` : ``
...
+using Pkg
+Pkg.rebuild("Conda")
```
stevengj, line 201: build, not rebuild.

stevengj: Note that they will need to restart Julia before or after this (i.e. before they try using Conda again)

src/Conda.jl
Alexander-Barth: The idea would be that the user would just need to copy and paste the command. The error message would either print-out a double backslash or use `raw"C:\Conda-Julia\3"` (I use the latter in the updated PR).

Alexander-Barth:
Thank you very much for your feed-back. I just updated the PR. With the revised PR, the error message now reads:

```
ERROR: Conda.jl cannot be installed to its default location C:\Users\Clément\.julia\conda\3
as Miniconda does not support the installation to a directory with a space or a
non-ASCII character on Windows. The work-around is to install Miniconda to a
user-writable directory by setting the CONDA_JL_HOME environment variable. For
example on Windows:

ENV["CONDA_JL_HOME"] = raw"C:\Conda-Julia\3"
using Pkg
Pkg.build("Conda")

The Julia session need to be restarted. More information is available at
https://github.com/JuliaPy/Conda.jl.
```

Alexander-Barth:
I am wondering if this PR can be merged now. Should I rebase it?

mkitti:
Hmm, if you rebase I will merge.

Alexander-Barth:
I just rebased the PR. I now see this failure on windows:

```
[ Info: Running `conda create -q -y -p 'C:\Users\runneradmin\.julia\conda\3\x86_64\envs\test_conda_jl'` in root environment
...
 File "C:\Users\runneradmin\.julia\conda\3\x86_64\lib\site-packages\urllib3\_request_methods.py", line 10, in <module>
    from .response import BaseHTTPResponse
  File "C:\Users\runneradmin\.julia\conda\3\x86_64\lib\site-packages\urllib3\response.py", line 282, in <module>
    class BaseHTTPResponse(io.IOBase):
  File "C:\Users\runneradmin\.julia\conda\3\x86_64\lib\site-packages\urllib3\response.py", line 292, in BaseHTTPResponse
    DECODER_ERROR_CLASSES += (brotli.error,)
AttributeError: module 'brotli' has no attribute 'error'
```

This seems to be a known conda issue:
https://github.com/conda/conda/issues/9903

stevengj (approved):

mkitti:
I encountered the same issue earlier.

PR #225, opened 2022-09-27
remove `--source-cache` from conda clean

t-bltg:
`conda clean --source-cache` was deprecated in https://github.com/conda/conda/pull/7731, and now errors.


t-bltg:
Dupe of https://github.com/JuliaPy/Conda.jl/pull/219.

PR #232, opened 2023-02-06
include Sys.ARCH in default rootenv

stevengj:
This is becoming important on e.g. MacOS where people commonly run both the ARM and x86_64 architectures on the same machine (the latter via Rosetta).


stevengj:
Unrelated CI failure `UnavailableInvalidChannel: HTTP 404 NOT FOUND for channel  foo`

PR #233, opened 2023-02-07
fix tests?

stevengj:
It seems that Conda no longer likes having a fake channel "foo"…


stevengj:
No, now getting `HTTP 404 NOT FOUND for channel  defaults `?

PR #241, opened 2023-06-12
Add satified_skip_solve as a keyword argument for Conda.add

mkitti:
Add --satisified-skip-solve as an argument for Conda.add.

This will allow PyCall.jl to ensure that numpy is installed, but not update it if it already in the environment.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -268,6 +268,8 @@ end
         end
     end
 
+    #=
+    # This is broken
     @testset "version mismatch" begin
         preserve_build() do
             # Mismatch in written file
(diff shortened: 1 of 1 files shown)


MilesCranmer:
LGTM

The CI is broken FYI

MilesCranmer:
Broken since November: https://github.com/JuliaPy/Conda.jl/commits/master

mkitti:
Yes, but why is CI so broken.

MilesCranmer:
@stevengj any idea?

mkitti:
What is happening here?
AttributeError: module 'brotli' has no attribute 'error'

stevengj:
LGTM.  The CI problem on Windows is not a regression so it shouldn't be blocking.

davidanthoff:
This seems to breaks CI on x86 ubuntu workflows, see [here](https://github.com/queryverse/ExcelReaders.jl/actions/runs/5563851312/jobs/10163073901?pr=101#step:4:117).

Could this be reverted for now?

Also, the CI here seems unfortunately limited, shouldn't it run on more Julia versions and on things like x86?

mkitti:
I guess that is because there is no miniforge for x86:
https://github.com/conda-forge/miniforge

mkitti:
Is there commercial miniconda for x86? https://docs.conda.io/en/latest/miniconda.html

davidanthoff:
There seems to be _something_, right? It does look like some kind of `conda` gets installed... There is an `info` that says that `miniconda` is being installed.

mkitti:
My guess is that we are pulling in a very old version of conda when x86 was supported. I will investigate further before trying to sort out the fjx.

mkitti:
@davidanthoff My proposed fix is https://github.com/JuliaPy/PyCall.jl/pull/1045

Issue #242, opened 2023-07-07
disable auto_update_conda?: conflict conda versions when installing RCall.jl and PyCall.jl


szcf-weiya:
I encountered the error in the AutoMege step when registering my package that depends on RCall.jl. Here is the pull request on the General repo and the action log info.

https://github.com/JuliaRegistries/General/pull/86838
https://github.com/JuliaRegistries/General/actions/runs/5464785321/jobs/9947294633?pr=86838

```
    Building Conda ─→ `/tmp/jl_aWu15K/scratchspaces/44cfe95a-1eb2-52ea-b672-e2afdf69b78f/915ebe6f0e7302693bdd8eac985797dba1d25662/build.log`
    Building PyCall → `/tmp/jl_aWu15K/scratchspaces/44cfe95a-1eb2-52ea-b672-e2afdf69b78f/43d304ac6f0354755f1d60730ece8c499980f7ba/build.log`
    Building RCall ─→ `/tmp/jl_aWu15K/scratchspaces/44cfe95a-1eb2-52ea-b672-e2afdf69b78f/d441bdeea943f8e8f293e0e3a78fe2d7c3aa24e6/build.log`
ERROR: Error building `RCall`: 
Collecting package metadata (current_repodata.json): ...working... done
Solving environment: ...working... unsuccessful initial attempt using frozen solve. Retrying with flexible solve.
Solving environment: ...working... unsuccessful attempt using repodata from current_repodata.json, retrying with next repodata source.

ResolvePackageNotFound: 
  - conda==23.1.0

[ Info: Installing R via Conda.  To use a different R installation, set the "R_HOME" environment variable and re-run Pkg.build("RCall").
[ Info: Running `conda config --add channels r --file /tmp/jl_aWu15K/conda/3/x86_64/condarc-julia.yml --force` in root environment
[ Info: Running `conda install -y 'r-base>=3.4.0,<4'` in root environment
[...]

Issue #238, opened 2023-05-20
ResolvePackageNotFound:    - conda==23.1.0


schlichtanders:
Hi, I am using Conda.jl and it failed on the most recent Julia 1.9.0 docker image to install packages

I explicitly set the environment variable `PYTHON=""` to force using Conda.jl default conda installation.
You can for instance start the docker with
```bash
docker run -it --rm julia:1.9 bash
``` 
then starting julia in this bash console you can run:
```julia
ENV["PYTHON"]=""
using Conda
Conda.add(["numpy", "pandas", "matplotlib", "plotly"])
```
Which should output someting like
```
[ Info: Running `conda install -q -y numpy pandas matplotlib plotly` in root environment
Collecting package metadata (current_repodata.json): ...working... done
Solving environment: ...working... failed with initial frozen solve. Retrying with flexible solve.
Solving environment: ...working... failed with repodata from current_repodata.json, will retry with next repodata source.

ResolvePackageNotFound: 
  - conda==23.1.0

ERROR: LoadError: failed process: Process(setenv(`/github/home/.julia/conda/3/x86_64/bin/conda install -q -y numpy pandas matplotlib plotly`,["PATH=/usr/local/julia/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin", "JOLIN_ENVIRONMENT=test", "GITHUB_RUN_NUMBER=16", "JULIA_PATH=/usr/local/julia", "GITHUB_REF_NAME=45/merge", "RUNNER_ARCH=X64", "GITHUB_REF=refs/pull/45/merge", "GITHUB_STEP_SUMMARY=/__w/_temp/_runner_file_commands/step_summary_ad717c8d-77e6-488e-8547-c929ccb04ec2", "GITHUB_REPOSITORY=jolin-io/JolinWorkspaceTemplate", "TZ=UTC\u2060"  …  "GITHUB_WORKFLOW=Test", "GITHUB_ACTION_REPOSITORY=", "GITHUB_REPOSITORY_OWNER=jolin-io", "GITHUB_EVENT_NAME=pull_request", "HOME=/github/home", "GITHUB_ACTION=__run_4", "OPENBLAS_MAIN_FREE=1", "HOSTNAME=0051195642c0", "PYTHONIOENCODING=UTF-8", "GITHUB_EVENT_PATH=/github/workflow/event.json"]), ProcessExited(1)) [1]

Stacktrace:
[...]

PR #243, opened 2023-07-09
disable auto_update_conda

szcf-weiya:
The motivations come from #242 and #238, where the error "ResolvePackageNotFound:   - conda==23.1.0" is thrown when conda automatically upgrades from 23.1.0 to 23.5.0. 

Although I did not quite understand the real logic of the error, one simple workaround is to disable the auto-update of conda, i.e., fix its version.

On the other hand, since the package Conda.jl has already required to install the latest conda release, 

https://github.com/JuliaPy/Conda.jl/blob/fe6c94c70609e6b6a8f853d72efc7d890df967e6/src/Conda.jl#L167-L179

I think the later update by `auto_update_conda` is just minor, like from 23.1.0 to 23.5.0, then disabling such minor aggressive updates should be reasonable. As a byproduct, it can save the upgrade time.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -103,6 +103,8 @@ const CONDA_EXE = "$(escape_string(CONDA_EXE))"
 mkpath(condadir)
 mkpath(ROOTENV)
 
+write("$ROOTENV/condarc-julia.yml", "auto_update_conda: false")
+
 for depsfile in ("deps.jl", condadeps)
     if !isfile(depsfile) || read(depsfile, String) != deps
         write(depsfile, deps)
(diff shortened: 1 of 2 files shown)


mkitti:
Why are the Windows tests failing?

szcf-weiya:
I think it is just a simple edit, and should not be related to the platform.

How can we rerun the test? Because I noticed the near commits also failed on Windows.

image

stevengj:
Probably Conda fixed their packages in the meantime?  But I still feel like it is good practice not to upgrade unless the user specifically requests `Conda.update()`.

PR #244, opened 2023-07-15
Test on LTS and x86

davidanthoff:


stevengj:
The CI failures here are not regressions, I guess?

mkitti:
There is no miniforge support for x86:
https://github.com/conda-forge/miniforge

The last support for miniconda is Python 3.9:
https://docs.conda.io/en/latest/miniconda.html

Python 3.9 is supported through October 2025:
https://devguide.python.org/versions/

mkitti:
Could we merge the 1.6 (LTS) testing without x86?

ViralBShah:
Trying to simplify to a smaller set and refresh the CI script here. Hoping to merge this and then get more platforms etc. in later.

ViralBShah:
#259 should hopefully pull in a new enough version of numpy that it is available on arm macs.

PR #246, opened 2023-07-19
Issue warning and then ignore satisified_skip_solve on 32-bit windows

mkitti:
Since conda has stopped being updated for 32-bit Windows and does not have `--satisfied-skip-solve`, issue a warning that it is not supported on that platform and then ignore the keyword on `Conda.add`.


stevengj:
LGTM.

stevengj (approved):

mkitti:
Merging expeditiously to resolve breakage then

PR #262, opened 2025-10-03
Add support for new JSON 1.0 release

### LINKED ISSUE #264: JSON is restricted to version 0.21
The package JSON is restricted to version at most 0.21. There was a major release of JSON 1.0.1. Could you upgrade the Project.toml to allow using this new version?

quinnj:
Fixes #264

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -26,7 +26,7 @@ jobs:
             arch: aarch64
             version: '1'
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
       - uses: julia-actions/setup-julia@v2
         with:
           version: ${{ matrix.version }}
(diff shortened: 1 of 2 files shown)


quinnj:
I think the failures here are unrelated? Something about checking pip interop?

lgoettgens:
It would be great to get this in. We converted our packages to JSON 1.0 (it unfortunately was not possible to allow 0.21 at the same time), and now we are unable to use the dev version of the package inside Jupyter notebooks.

@ViralBShah @stevengj, from the commit history, it looks as if you two are mostly maintaining this. Would it be possible if one of you has a look at this? Thanks!

hhaensel:
just filed a PR in IJulia https://github.com/JuliaLang/IJulia.jl/pull/1197 that works with JSON v1 as soon as Conda's json compat is bumped to include 1

